PR #255, opened 2026-06-06
chore: improve EverOS maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in tests/conftest.py, tests/e2e/conftest.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -51,4 +51,13 @@ def _reset_settings_cache() -> Iterator[None]:
 @pytest.fixture(scope="session")
 def long_conversation() -> dict:
     """LoCoMo conv_0 fixture (419 messages, 19 batches, one session)."""
-    return json.loads(_LONG_CONV_PATH.read_text())
+    data: dict = json.loads(_LONG_CONV_PATH.read_text())
+    assert isinstance(data, dict)
+    assert "everos_session_id" in data
+    assert "batches" in data
+    batches: list = data["batches"]
+    assert isinstance(batches, list)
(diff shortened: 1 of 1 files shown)


cyfyifanchen:
Hi @lphuc2250gma, thank you for contributing to EverOS.

I checked this against the current EverOS 1.0.0 codebase. The change is small and technically compatible, but we are not going to merge this PR as-is because the validation is added inside the shared `long_conversation` fixture rather than as a focused fixture-integrity test. That means the checks only run as a side effect of tests that happen to load the fixture. The assertions also use fallback defaults for `total_batches` and `total_messages`, which weakens the validation if those metadata fields are accidentally removed.

For a follow-up contribution, the cleaner shape would be:

1. Open a fresh branch from the current `main`.
2. Add a dedicated fixture integrity test, for example under `tests/unit/` or a focused fixture test module.
3. Validate `everos_session_id`, `batches`, `total_batches`, `total_messages`, and each batch's `message_count == len(messages)` without fallback defaults.
4. Use a specific title such as `test: validate long conversation fixture shape`.
5. Make sure all required checks pass before review.

I am closing this PR to keep the queue focused after the EverOS 1.0.0 reset, but the direction is useful and very welcome. Looking forward to seeing more contributions like this. Thank you so much.

PR #284, opened 2026-06-11
fix: ignore tool-call items when collecting user senders

### LINKED ISSUE #276: User-side extraction strategies crash on ToolCallRequest because they assume every memcell item has role
## Summary

Some user-side extraction strategies assume every item in `memcell.items` has a `role` attribute. Tool-call trajectories can include `ToolCallRequest` items, which do not have `role`, causing downstream strategies to crash and eventually enter `dead_letter`.

Observed affected strategies:

```text
extract_atomic_facts
extract_foresight
```

## Why this matters

Tool-use trajectories are common for coding agents and assistant workflows. A single tool-call item should not break user memory extraction for the whole memcell.

In the observed run, the agent case itself was extracted, but user-side extraction strategies failed independently because they iterated over mixed item types as if every item were a normal conversational message.

## Observed evidence

OME records:

```text
extract_atomic_facts: dead_letter
extract_foresight: dead_letter
```

Failure message:

```text
```

Relevant code locations:

```text
src/everos/memory/strategies/extract_atomic_facts.py:66
src/everos/memory/strategies/extract_foresight.py:54
src/everos/service/_boundary.py
```

The extraction strategies contain logic shaped like:

```python
```

But `_boundary.py` can map assistant tool calls into `ToolCallRequest`, which does not expose `role`.

## Reproduction shape

Use a synthetic agent/coding trajectory with assistant tool calls, for example:

1. User reports a failing focused test.
2. Assistant emits a tool call to run the test.
3. Tool returns trace output.
4. Assistant identifies the root cause.
5. Assistant emits another tool call to apply or validate a fix.
6. Tool returns passing output.
7. Assistant summarizes the result.

Then flush/process the session. The memcell contains mixed item types, including tool-call request objects.

## Expected behavior

Tool-call items should not crash extraction strategies that only need user messages.

Possible fixes:

- Filter by item type before reading `role`.
- Use `getattr(m, "role", None) == "user"` where appropriate.
- Centralize sender/user-message extraction in a helper that understands mixed memcell item types.
- Ignore non-text/tool-call items in user-side strategies unless they are explicitly supported.

## Environment

```text
EverOS: 1.0.0 source checkout
Runtime: Docker Linux runtime
Python: 3.12
Data: synthetic agent trajectory only
```

No real user memory or secrets were used.

rendigua2025-gif:
## Summary

Fixes #276.

User-side extraction strategies assumed every item in `memcell.items` exposes a `role` attribute. Agent/tool-call trajectories can include `ToolCallRequest` items, which do not expose `role`, so `extract_atomic_facts` and `extract_foresight` could crash before processing otherwise valid user messages.

This PR adds a small shared helper for collecting user sender ids defensively, then uses it in both strategies.

## Scope

- Collect sender ids only from items where `role == "user"`.
- Skip non-chat/tool-call items instead of treating them as malformed messages.
- Preserve the existing extraction behavior for normal chat messages.

This does not add tool-call semantics to user-side extraction; it only prevents unrelated tool-call items from crashing the user memory pipeline.

## Tests

- `tests/unit/test_memory/test_strategies/test_strategy_to_handler_contract.py`
- `tests/unit/test_memory/test_strategies/test_extract_atomic_facts.py`
- `tests/unit/test_memory/test_strategies/test_extract_foresight.py`

Note: I verified locally on Windows with a test-process-only `fcntl` stub because the current repository still imports POSIX `fcntl` during test collection on Windows.

### DIFF
--- a/src/everos/memory/strategies/_sender_utils.py
+++ b/src/everos/memory/strategies/_sender_utils.py
@@ -1,0 +1,24 @@
+"""Shared sender-collection helpers for strategy modules."""
+
+from __future__ import annotations
+
+from everalgo.types import MemCell
+
+
+def collect_user_sender_ids(memcell: MemCell) -> list[str]:
+    """Distinct ``role='user'`` sender_ids in stable sorted order.
+
+    User-side strategies may receive agent-trajectory cells that include
+    ``ToolCallRequest`` items alongside chat messages. Those items expose
+    ``sender_id`` but not ``role``, so sender discovery must probe the
+    attribute defensively instead of assuming every item is a ChatMessage.
+    """
+
+    sender_ids: set[str] = set()
+    for item in memcell.items:
+        if getattr(item, "role", None) != "user":
+            continue
+        sid = getattr(item, "sender_id", None)
+        if sid:
+            sender_ids.add(sid)
+    return sorted(sender_ids)
--- a/src/everos/memory/strategies/extract_atomic_facts.py
+++ b/src/everos/memory/strategies/extract_atomic_facts.py
@@ -39,6 +39,7 @@ from everos.infra.ome.triggers import Immediate
 from everos.infra.persistence.markdown import AtomicFactWriter
 from everos.memory.events import UserPipelineStarted
 from everos.memory.models import AtomicFact
+from everos.memory.strategies._sender_utils import collect_user_sender_ids
 
 logger = get_logger(__name__)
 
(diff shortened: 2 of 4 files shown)


cyfyifanchen:
Thanks for catching this and putting together the defensive sender helper. The affected paths have since changed: atomic-fact extraction now receives a single-owner episode event, and foresight filters for chat messages before reading role.

The foresight fix and mixed tool-call regression coverage landed with #393 in v1.2.3: https://github.com/EverMind-AI/EverOS/pull/393

That covers the crash this PR addresses, so I'm closing the older patch as superseded, along with #276. We appreciate the work you put into it.

Issue #275, opened 2026-06-09
Agent skill extraction can dead-letter when SkillClusterUpdated fires before agent_case is visible in LanceDB


rendigua2025-gif:
## Summary

`extract_agent_skill` can enter `dead_letter` when `SkillClusterUpdated` is emitted before the newly written `agent_case` is visible through LanceDB.

In an isolated Linux/Docker run with a synthetic agent trajectory, EverOS successfully generated and indexed an `agent_case`, but the downstream `extract_agent_skill` strategy retried too early and exhausted its retries with `_CaseNotYetIndexedError`.

## Why this matters

Agent memory is one of EverOS's differentiating capabilities. In this case:

- `extract_agent_case` succeeded.
- `trigger_skill_clustering` succeeded.
- `agent_case` Markdown existed.
- `agent_case` was later visible in LanceDB.
- `/search` could retrieve the `agent_case`.
- But `extract_agent_skill` still ended in `dead_letter` because it checked LanceDB before the case was visible.

That means the case-to-skill chain can fail even though the case itself is valid and eventually searchable.

## Reproduction shape

Synthetic agent trajectory:

1. User reports a focused test failure.
2. Assistant/agent runs a focused test.
3. Tool returns a trace.
4. Assistant identifies root cause.
5. Assistant patches the minimal code path.
6. Tool reports focused test pass.
7. Assistant runs the broader test module.
8. Assistant summarizes the reusable debugging lesson.

The sample was sufficient to generate an `agent_case` with quality score `0.98`.

## Observed evidence

`agent_case` result:

```text
agent_case: 1
agent_skill: 0
agent_case search: keyword/vector/hybrid all retrieved the case
```

OME records:

```text
extract_agent_case: success
trigger_skill_clustering: success
extract_agent_skill: dead_letter
```

Failure message:

```text
_CaseNotYetIndexedError: AgentCase entry_id=ac_20260609_00000001 not in LanceDB yet; retrying
```

Relevant code locations:

```text
src/everos/memory/strategies/trigger_skill_clustering.py
src/everos/memory/strategies/extract_agent_skill.py
```

`trigger_skill_clustering` emits `SkillClusterUpdated`, and `extract_agent_skill` then calls `_load_target_case`, which raises `_CaseNotYetIndexedError` if the case is not yet visible in LanceDB.

## Expected behavior

One of the following would make the chain more reliable:

[...]

PR #286, opened 2026-06-11
fix: wait for agent_case LanceDB visibility before skill extraction

rendigua2025-gif:
## Summary

Addresses #275.

`SkillClusterUpdated` can arrive before the corresponding `agent_case` is visible through LanceDB. Previously, `extract_agent_skill` checked LanceDB once, raised `_CaseNotYetIndexedError`, and could quickly consume OME retry budget during normal cascade/indexer lag.

This PR adds a short bounded visibility wait inside `_load_target_case` before raising the existing retry-class error.

## Scope

- Poll `agent_case_repo.find_by_owner_entry(...)` for a small bounded window.
- Return immediately once the target case becomes visible.
- Preserve the existing `_CaseNotYetIndexedError` path when the case is still not visible after the window.

This is a low-risk mitigation, not a full architectural fix. If LanceDB visibility lag exceeds the short wait window, OME retry still handles the delayed dependency. A deeper fix could emit `SkillClusterUpdated` only after confirmed case index visibility, carry enough case data in the event payload, or model dependency-not-ready as delayed scheduling.

## Tests

- `tests/unit/test_memory/test_strategies/test_extract_agent_skill.py`
- `tests/integration/test_ome_strategies_integration.py -k test_skill_chain_e2e`

Note: I verified locally on Windows with a test-process-only `fcntl` stub because the current repository still imports POSIX `fcntl` during test collection on Windows.

### DIFF
--- a/src/everos/memory/strategies/extract_agent_skill.py
+++ b/src/everos/memory/strategies/extract_agent_skill.py
@@ -34,6 +34,8 @@ short-circuits low-quality cases internally via its own
 
 from __future__ import annotations
 
+import asyncio
+
 from everalgo.agent_memory import AgentSkillExtractor
 from everalgo.types import AgentCase as AlgoAgentCase
 from everalgo.types import AgentSkill as AlgoAgentSkill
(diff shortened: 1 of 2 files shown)


cyfyifanchen:
Thanks for investigating this race and proposing a bounded wait. The target case now travels in the event payload, so skill extraction no longer waits for that case to become visible in the index.

This landed in #393 and shipped in v1.2.3: https://github.com/EverMind-AI/EverOS/pull/393

I'm closing this polling patch as superseded by that change, along with the original issue #275. Thanks for taking the time to trace where the case-to-skill chain could break.

PR #291, opened 2026-06-16
docs: add local agent smoke runbook

UncleJ-h:
## Summary
- Add a local agent smoke runbook for agent builders integrating EverOS with multiple local runtimes.
- Cover loopback proxy bypass, write + flush, cascade sync for read-your-write, cross-runtime read, source markdown deletion, and /get + /search absence checks.
- Link the new guide from the documentation index.

## Why
Agent runtimes often run in shells with proxy variables, separate process state, and automatic hook ambitions. This runbook gives builders a small manual proof before they wire EverOS into production agent hooks.

## Validation
- `git diff --check`
- Verified linked doc targets exist locally
- Cross-checked the runbook claims against existing `api.md`, `how-memory-works.md`, `storage_layout.md`, `cli.md`, and `SECURITY.md`

## Scope
Docs only. No runtime behavior changes.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -37,6 +37,7 @@ specific thing (drain a queue, recover from a stuck row, etc.).
 | Doc | Purpose |
 |---|---|
 | [cascade_runbook.md](cascade_runbook.md) | Cascade subsystem ops — drain queue, recover stuck rows |
+| [local-agent-smoke.md](local-agent-smoke.md) | Local agent integration smoke — write, cross-runtime read, delete, and absence proof |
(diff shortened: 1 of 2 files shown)


UncleJ-h:
Addressed Copilot review feedback in 01f6f77.\n\n- Replaced the mktemp-based smoke root command with python3 tempfile.mkdtemp() to avoid GNU/BSD mktemp differences.\n- Replaced the rm glob deletion with find ... -print -delete, and added a note to stop if no episode file is printed.\n- Verified git diff --check, tempfile creation, empty-match delete behavior, and existing episode deletion locally.

PR #317, opened 2026-06-30
fix(memory): skip ToolCallRequest items in foresight sender scan

### LINKED ISSUE #276: User-side extraction strategies crash on ToolCallRequest because they assume every memcell item has role

Ayush7614:
## Summary

- Fix `extract_foresight` crashing on agent trajectories that include `ToolCallRequest` items (which have no `role` attribute).
- Use `getattr(item, "role", None)` when collecting user senders, matching the pattern already used in `_unique_user_senders` and `_collect_agent_ids`.
- Add a unit test with a mixed memcell (user message + tool call + assistant reply).

Closes #276.

Note: `extract_atomic_facts` no longer iterates `memcell.items` on current `main` (it extracts from episode text), so the reported crash there appears already addressed by a prior refactor.

## Test plan

- [x] `uv run pytest tests/unit/test_memory/test_strategies/test_extract_foresight.py -q`
- [x] `make ci` (lint + unit + integration + package build)

### DIFF
--- a/src/everos/memory/strategies/extract_foresight.py
+++ b/src/everos/memory/strategies/extract_foresight.py
@@ -51,7 +51,13 @@ def _get_writer() -> ForesightWriter:
 async def extract_foresight(event: UserPipelineStarted, ctx: StrategyContext) -> None:
     # 1. List the user senders in this memcell.
     memcell = event.memcell
-    sender_ids = sorted({m.sender_id for m in memcell.items if m.role == "user"})
+    sender_ids = sorted(
+        {
+            m.sender_id
+            for m in memcell.items
+            if getattr(m, "role", None) == "user" and m.sender_id
+        }
+    )
     extractor = ForesightExtractor(llm=get_llm_client()) if sender_ids else None
 
     # 2. Run the LLM extractor once per sender (prompt is per-sender).
--- a/tests/unit/test_memory/test_strategies/test_extract_foresight.py
+++ b/tests/unit/test_memory/test_strategies/test_extract_foresight.py
@@ -5,7 +5,14 @@ from unittest.mock import AsyncMock, patch
 
 import pytest
 import structlog.testing
-from everalgo.types import ChatMessage, Foresight, MemCell
+from everalgo.types import (
+    ChatMessage,
+    Foresight,
+    MemCell,
+    ToolCall,
+    ToolCallFunction,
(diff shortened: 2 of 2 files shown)


Ayush7614:
Closing — will not contribute on issues opened by others.

PR #339, opened 2026-07-13
docs(examples): add Langfuse (OpenTelemetry) integration example

dani1005:
Adds `examples/langfuse/` — a thin OpenTelemetry wrapper that traces EverOS memory operations (add / flush+extract / search / reflection) into Langfuse, with recall quality pushed as Langfuse scores.

- Pure OTel SDK, no Langfuse package dependency; works with any OTLP backend.
- Runs with no server via a built-in mock, or against a real EverOS server (`EVEROS_BASE_URL`). Verified end-to-end (traces + scores render in Langfuse).
- Additive only, no changes to EverOS core.
- Referenced by the upcoming Langfuse docs integration cookbook.

Native, opt-in instrumentation in core is a planned follow-up.

### DIFF
--- a/examples/langfuse/.gitignore
+++ b/examples/langfuse/.gitignore
@@ -1,0 +1,3 @@
+spans.jsonl
+__pycache__/
+.venv/
--- a/examples/langfuse/README.md
+++ b/examples/langfuse/README.md
@@ -1,0 +1,64 @@
+# EverOS × Langfuse (OpenTelemetry)
+
+Trace EverOS memory operations — writes, LLM extraction, recall with quality
+scores, and reflection — into [Langfuse](https://langfuse.com) as OpenTelemetry
+spans, so an agent's memory layer becomes visible and evaluable next to the rest
+of its traces.
+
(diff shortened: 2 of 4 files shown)


dani1005:
Heads-up for reviewers on structure:

This introduces a new top-level `examples/` directory (it didn't exist before). The intent is for it to be the home for runnable integration / usage examples, one per subfolder — this PR adds `examples/langfuse/`, and future ones would sit alongside (e.g. `examples/<integration>/`).

Why `examples/` rather than alternatives:
- It's runnable sample code users copy/adapt, not a shipped module — so `examples/` fits better than `integrations/` (which implies a first-party, packaged integration).
- It's code + a runnable demo, so `examples/` reads more naturally than nesting under `docs/`.

This is additive only — no changes to existing code or core. Native, opt-in instrumentation inside core is a planned follow-up.

If you'd prefer a different home for this (e.g. `docs/integrations/` or `integrations/`), happy to rename before merge — just let me know.

PR #357, opened 2026-07-24
fix(search): agent-agentic 500 + regenerate stale e2e seed/tests

gloryfromca:
## What

Two related fixes to the search subsystem's e2e coverage, split into two commits.

### 1. `fix(search)` — agent-agentic 500

`POST /api/v*/memory/search` with `owner_type=agent` + `method=agentic` returned **500**. The agent AGENTIC path fed recall candidates straight into everalgo's `aagentic_retrieve`, whose `_format_docs` (the LLM sufficiency / multi-query prompt) requires `metadata["episode"]` as a `{subject, content}` dict + a ms-epoch `timestamp`. Agent-kind rows carry their body in the recaller's `text_field` and time as a `datetime`, so `_format_docs` raised `TypeError: Candidate ... has no episode dict`.

Fix mirrors the episode path's bridge: reshape agent candidate metadata into the everalgo doc contract before `aagentic_retrieve`, then revert it before DTO shaping. Other agent methods (keyword/vector/hybrid) and user-memory agentic were unaffected.

### 2. `test(search)` — regenerate stale seed + migrate e2e

The committed `search_seed` fixture and several search e2e tests were written for the **pre-1.5 memcell** fact-linkage model. Current extraction links `atomic_fact.parent_id == episode.entry_id` (`parent_type="episode"`), and `user_memory` clusters store episode `entry_id` members. The stale fixture made VECTOR/AGENTIC recall + cluster narrowing find nothing.

- Regenerated `search_seed/*` from a fresh corpus in current entry_id format; facts bridge across multiple episodes.
- Fixed `_dump_search_seed.py` sampling (pick fact-hosting episodes first; keep facts by entry_id) so re-dumps stay coherent.
- Migrated e2e tests to the entry_id model (hierarchical-eviction, session/timestamp filters, cluster-seed helper); updated the filter-error assertion to the current `INVALID_INPUT` code.
- Provisioned `ome.toml` in the full-app pipeline fixture (OME config reloader requires it; strategies are code-registered so the packaged default suffices), unblocking corpus regeneration.

## Verification

- **Full search e2e suite: 49/49 green** (live LLM/rerank/embedder).
- `make lint` (ruff + import-linter + datetime + openapi), unit (1569), integration (79) all green.

## Notes

- These e2e are CI-excluded (`-m 'not slow and not live_llm'`); this PR restores them to green for local/manual runs. No product behavior change beyond the agent-agentic 500 fix.
- Independent of #356 (v2 API alias).

[...]

PR #354, opened 2026-07-24
fix(lancedb): detect schema type drift and add cascade rebuild recovery

gloryfromca:
## Problem — the root cause behind #337

`verify_business_schemas` (the startup guard) only compared column **names** against the current schema. A column whose on-disk Arrow **type** had drifted — name unchanged — sailed through and detonated much later, deep inside `merge_insert`, as the notoriously opaque:

```
RuntimeError: lance error: LanceError(IO): Execution error: Spill has sent an error, poll.rs:291:44
```

This is exactly #337: an `episode.subject_vector` column left as `string` (or `null`) by an older build, while the current schema declares a 1024-d `fixed_size_list`. Writing a real vector into that column is a pyarrow type error that lance wraps as the "Spill" IO error.

### Verified

- **Reproduced byte-identically**: a drifted `subject_vector` (`string` and `null`) column + a real 1024-d vector through the real `LanceRepoBase.upsert()` path yields the exact #337 error; a correctly-typed table accepts the same records.
- **No lancedb version produces the bad column** (checked 0.13→0.34: ≤0.19 raise `TypeError`, 0.20+ render the correct `fixed_size_list`), and lance never silently adds a mistyped column — so the drift originates from a build that declared the field differently, and the guard is what should have caught it at startup.

## Changes

**1. Type-aware `verify_business_schemas`** — compare each shared column's Arrow **type** against `schema.to_arrow_schema()` (the exact schema `get_table` builds tables from, so a healthy table never false-positives). On drift the startup error now names the drifted column and both types instead of the runtime Spill.

**2. `everos cascade rebuild`** — the safe recovery from a drifted/corrupt index:
- drops the business LanceDB tables and re-indexes from markdown (the source of truth);
- **skips** the verify guard (whose whole point is the drift it would otherwise trip on at startup — chicken-and-egg);
- unlike `rm -rf ~/.everos/.index/lancedb`, **re-populates already-`done` entries** (`reset_all` clears the cascade queue so every md file re-enqueues) — a bare `rm` of the lancedb dir leaves the queue marked done and the index comes back empty;
- unlike `rm -rf ~/.everos/.index`, **preserves `unprocessed_buffer`** (messages received but not yet extracted — not rebuildable from md).

## Tests

[...]

PR #379, opened 2026-07-31
fix(lancedb): reclaim stale versions via write-locked prune

gloryfromca:
Consolidated LanceDB hardening PR. Folds in #354 / #355 / #357 (now closed) plus the prune-under-lock storage fix + cascade health, all rebased on latest `main`. This mirrors the code the endurance soak actually exercised.

## Commits

- `fix(deps)`: pin lancedb to `>=0.34.0,<0.35.0` (was #355) — 0.35 embeds lance-rust v9; pin the version the soak validated.
- `fix(search)`: agent-agentic 500 + regenerate stale e2e seed/tests (was #357).
- `fix(lancedb)`: detect schema type drift + `cascade rebuild` recovery (was #354).
- `fix(lancedb): reclaim stale versions via write-locked prune` — the main change (below).
- `chore(rebase)`: adapt #354's integration test to main's soft-embedding `CascadeOrchestrator`.

## The prune-under-lock fix (why)

A storage-endurance soak (~100h across three runs, sustained churn + fault injection) reproduced the disk-bloat root cause: the bundled lock-free `optimize(cleanup_older_than=...)` loses its commit-conflict race against concurrent writes, so cleanup ran only **16 of ~250 times** and superseded dataset versions piled up until the index dir hit the 40G disk guardrail and never reclaimed under load.

Split the maintenance path:

- **`LanceRepoBase.optimize()`** — compact-only, lock-free (a `Retryable commit conflict` here is benign; the next beat retries, so it must not stall writers).
- **`LanceRepoBase.prune(older_than)`** — new; `cleanup_older_than + delete_unverified=False` **under the per-table write lock** (no writer in flight in-process → the Rewrite owns the manifest → cleanup completes every beat). `delete_unverified=False` keeps it safe against a *second process* (CLI `cascade sync` / `backfill`) — the in-process lock can't fence those, and `False` never deletes a file another writer may be mid-commit on; measured to reclaim identically to `True` on churned tables (~97% version collapse). Also removes empty `_indices/<uuid>/` husks (soak: 13061 dirs, 98% empty), threaded.
- Cascade worker: heavy beat calls `prune()`, light beat `optimize()`. Benign light-beat conflict → debug, doesn't count toward the failure streak or trigger a fallback rebuild.
- Prune retention decoupled from the prune cadence, defaulted to **60s** (runs under the lock → only needs to outlive an in-flight read).

Plus cascade **observability**: `CascadeOrchestrator.health()` → `CascadeHealth`, surfaced as a typed `cascade` block on `GET /health`. `healthy` = operational only (drain / optimize / prune); `failed_permanent` is an informational count that does not flip `healthy`.

## Soak result

- Disk **bounded & fully reclaimable** — sawtooths under load, reclaims to live size (~1.3G); 42G → 1.3G in 10 min once writes pause. vs run1 stuck at 40G until idle.
- 0 crashes / 0 OOM / 0 stuck cleanups across all runs. lance reports 2 live fragments / ~625MB per table.

Full write-up (code + conditions + results, with figures): DlabDev wiki → "EverOS LanceDB 存储浸泡测试 — 最终报告".

## Tests

`make lint` + import-linter + `dump_openapi.py --check` + full unit suite (**1882 passed**).

Note: `tests/e2e/test_search_endpoint_e2e.py` changes one assertion from `HTTP_ERROR` to `INVALID_INPUT`. That is a migration of a pre-existing slow-marked test already drifted on `main` (`exception_handlers.py` returns `INVALID_INPUT` for this path) — this PR does **not** change any API error codes.

## Follow-up (not in this PR)

Embedding/vector errors during cascade indexing are currently classified permanent (`retryable=False`); the soak's fuzz showed a transient embedder blip permanently fails a file. Candidate: reclassify as `RecoverableError`.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


gloryfromca:
Thanks for the thorough review — every finding was verified against the code and addressed. Resolution summary:

## P0

**P0-1 — backfill silently lost compaction/cleanup (masked by a stale test double)**
`_backfill.py` still called the removed `optimize(cleanup_older_than=…)`, which raised `TypeError` swallowed by the best-effort `try/except`, so `cascade backfill` skipped all reclaim — the exact bloat this PR fixes. CI stayed green because `_FakeRepo` kept the old signature. Fixed the call site to `optimize()` + `prune(timedelta(0))`; the fake now mirrors the real `optimize()`/`prune()` signatures (the drift that hid it), and the backfill tests pin the split. (`df142c0`)

**P0-2 — `delete_unverified=True` guarded only by an in-process lock**
Switched prune to `delete_unverified=False`. Measured on churned tables, `False` reclaims identically to `True` (both collapse superseded versions ~97%); `True` only additionally deletes in-flight/dangling files — exactly the cross-process corruption vector. So no cross-process lock is needed and the write-lock/commit fix (the real reclaim win) is unchanged. Runbook updated. (`df142c0`)

## P1

**P1-3 — `/health` could 500 on a SQLite hiccup**
Wrapped `orch.health()` in `try/except`: on failure it reports `healthy=false` + a reason and keeps HTTP 200, so a locked/full/migrating DB can't flip liveness and restart the container. Added a probe-exception test. (`df142c0`)

**P1-4 — `cascade rebuild` had no "stop the server" guard**
Runbook now states rebuild requires stopping the server (it drops+recreates tables; a live daemon holds cached handles pointing at the dropped dataset). The `cascade sync` "safe in parallel" note is reworded to explain *why* it's safe (`delete_unverified=False`). (`df142c0`)

## P2
- `_is_benign_commit_conflict` narrowed to the `"commit conflict"` phrase (dropped the over-broad `"retryable"` that would swallow unrelated recoverable errors).
- Added a timeout around the prune cleanup so a hung lance call can't wedge the per-table write lock.
- Corrected the stale `_run_optimize_once` / `_prune_stale_seconds` docstrings. (`df142c0`)

**Follow-up (`fb3dac7`)** — the schema-drift error's docstrings still described recovery as `rm -rf …/.index/lancedb`, which the runbook flags as the *wrong* recovery (it leaves the cascade queue `done`, so nothing re-indexes and the index comes back empty). Aligned them to `everos cascade rebuild`, and removed 15 dangling references to an internal numbered design-doc set that was never shipped to this repo.

## Validation — run5 (targeted, on `df142c0`)
A 3h concurrent run exercising what single-process endurance can't:
- **Cross-process integrity** — 375 concurrent `cascade sync` prunes alongside the daemon: **CLEAN** (schema verify passed, all tables reopen + count OK, md↔lance counts consistent). Only 4 transient retryable commit-conflicts across 3h, each self-healed (`optimize_failure_streak` never left 0, `/health` stayed healthy).
- **Disk bounded (`delete_unverified=False`)** — du sawtoothed to ~4G under churn, collapsed to 328MB on write-stop, then flat for 2+ days (zero drift).
- **P0-1 on real hardware** — `optimize()` + `prune(0)` on a live table, no `TypeError`.
- Zero OOM / crashes / add-write errors throughout.

One non-blocking observation from run5: a heavy-beat (prune) retryable commit-conflict is currently counted toward `optimize_failure_streak` (unlike the light beat). Under 2s-interval hammering it triggered only 4× in 3h and self-healed every time, never nearing the alert threshold — so it's a non-issue in practice, but treating the heavy-beat retryable conflict as benign too would be a clean optional follow-up. Happy to fold it in if you'd like.

Kendrick-Song (approved):

Issue #364, opened 2026-07-27
写入失败被上报为成功:cascade worker 不可恢复时 flush 仍返回 extracted;且失败无重试上限


arelchan:
## 一句话

cascade worker 处于持续失败状态时,写入 API 仍然返回成功 —— `POST /api/v1/memory/add` 返回 200、`POST /api/v1/memory/flush` 返回 `{"status": "extracted"}`,而实际没有任何内容落库。宿主应用无法从 API 契约上察觉记忆已经停止工作。同时,失败条目没有重试上限,单次运行可产生数百条完全相同的不可恢复日志。

这是从 #337 排查中分离出来的两个**独立**问题。#337 的具体成因(`subject_vector` schema 漂移)已由 #354 检测 + 恢复,但下面两点不随之解决。

### 复现

在一个 episode 写入必然失败的库上(例如 #337 的 schema 漂移状态):

```bash
curl -X POST localhost:1995/api/v1/memory/add -H 'Content-Type: application/json' \
  -d '{"user_id":"default","app_id":"default","session_id":"probe",
       "messages":[{"role":"user","sender_id":"default","timestamp":1769500000000,"content":"probe"},
                   {"role":"assistant","sender_id":"agent","timestamp":1769500000001,"content":"ack"}]}'
# -> 200 {"data":{"message_count":2,"status":"accumulated"}}

curl -X POST localhost:1995/api/v1/memory/flush -H 'Content-Type: application/json' \
  -d '{"user_id":"default","session_id":"probe"}'
# -> 200 {"data":{"status":"extracted"}}
```

同时服务端日志里是:

```
cascade_worker_unrecoverable kind=episode ...
RuntimeError: lance error: LanceError(IO): Execution error: Spill has sent an error
```

随后检索写入的内容,搜不到;`episode` 表行数不变。

### 影响

`"status": "extracted"` 在语义上表示抽取已完成,调用方据此认为写入成功。实际情况是这条记忆永久不会进入索引。

在我们的场景里,这导致 episode 写入**静默失败了 18 天**(表停在 7-09,期间零写入)才被发现 —— 因为:

- 写入侧 API 一直返回成功;
- recall 仍能召回旧数据(其他 kind 如 `atomic_fact` / `foresight` 写入正常),看起来"记忆还活着";
- 宿主(Raven)的记忆后端是 fail-open 的:捕获异常后返回空结果,不向上抛,所以它自己的链路追踪里这些节点也全是绿的。

[...]

PR #380, opened 2026-08-01
feat(api): expose cascade queue counters in /health (#364)

arkh-node:
## Summary

Closes part 1 of #364 — the half that makes the failure invisible rather than the half that causes it.

An HTTP host currently cannot tell that memory writes stopped reaching the index. Writes land in markdown and are indexed later by the cascade worker, so `/memory/flush` keeps answering `"extracted"` while the worker is stuck. In #364 that hid an 18-day episode outage: the write API returned success, recall still served older rows, and the fail-open host drew green traces over all of it.

I deliberately did **not** widen `flush`'s return type to `partial` / `degraded`, as the issue first suggested. `flush` isn't lying inside its own layer — it reports on the markdown write, which did succeed. Making it report on indexing would couple the write path to the cascade stage the architecture separates on purpose.

Instead: the counters already exist. `queue_summary()` computes `pending`, `done`, `failed_retryable`, `failed_permanent`, and `cascade status` prints all of them. They simply had no HTTP exit — a CLI operator could see the outage, an API host could not. That asymmetry is the defect this PR closes.

**What changed**

- `GET /health` gains an additive `cascade` section with the four existing counters plus `lag` (`max_lsn - last_processed_lsn`, the same figure the CLI derives).
- No existing field changes meaning, so the endpoint contract holds.
- When queue state is unreadable, `cascade` is `null` rather than zeroed. `/health` is a liveness probe and must answer even when the metadata store is down — and "unknown" must not read as "clean", which is precisely the confusion that let the original outage hide.

A host can now probe actively: `failed_permanent > 0` means rows are terminally stuck, no retry will clear them, and an operator has to run `cascade fix` / `cascade rebuild`.

Part 2 of the issue (no retry ceiling) is **not** addressed here. The `except Exception` branch already marks unrecoverable rows terminal via `mark_failed(retryable=False)`, so 330 identical log lines for the same entries point at re-enqueueing from elsewhere — most likely the scanner re-observing the same md files by mtime — rather than an unbounded retry loop. That deserves a reproduction before a patch, and it doesn't belong in the same change.

## Area

- [x] Developer experience
- [ ] Architecture method
- [ ] Benchmark
- [ ] Use case
- [ ] Documentation
- [ ] CI, build, or release

## Verification

```text
$ uv run pytest tests/unit/test_entrypoints/test_api/test_routes/test_health_cascade.py -v
test_cascade_counters_surface_terminal_failures PASSED
test_lag_never_goes_negative                    PASSED
test_unreadable_queue_degrades_to_null          PASSED
3 passed in 6.29s

$ uv run pytest tests/unit/test_entrypoints/ tests/integration/test_api/test_health_capabilities.py -q
220 passed in 102.84s

$ uv run ruff check src/everos/entrypoints/api/routes/health.py tests/.../test_health_cascade.py
All checks passed!

$ uv run ruff format --check <same files>
2 files already formatted
```

The new test pins three facts: terminal failures surface as `failed_permanent` with the LSN gap as `lag`; `lag` never goes negative; and an unreadable queue degrades to `null` instead of raising or reporting zeros.

Existing `/health` tests pass untouched.

### DIFF
--- a/src/everos/entrypoints/api/routes/health.py
+++ b/src/everos/entrypoints/api/routes/health.py
@@ -11,6 +11,10 @@ from everos.component.embedding import get_embedding_capability
 from everos.component.multimodal import get_multimodal_llm_capability
 from everos.component.parser import parser_available
 from everos.component.rerank import get_rerank_capability
+from everos.core.observability.logging import get_logger
+from everos.infra.persistence.sqlite import md_change_state_repo
+
+logger = get_logger(__name__)
 
 router = APIRouter(tags=["health"])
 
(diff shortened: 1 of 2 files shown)


cyfyifanchen:
Thanks for putting this together. The current /health endpoint now exposes cascade queue failures and operational readiness through the implementation added in #379, so we no longer need to merge this older version of the endpoint change.

There is still a gap worth keeping visible: the HTTP done and lag fields proposed here have not been implemented. They remain tracked in #364, which we're leaving open:
https://github.com/EverMind-AI/EverOS/issues/364#issuecomment-5150025428

Closing this PR as superseded, not marking the whole issue as resolved. Thanks for making the HTTP visibility gap concrete.

PR #430, opened 2026-09-02
fix(cascade): scope LanceDB row ids by md path to stop cross-project overwrites

### LINKED ISSUE #320: [Bug]: LanceDB row IDs collide across project_id scopes and overwrite searchable episodes
### Area

src/everos

## Summary

EverOS stores Markdown correctly under separate `project_id` directories, but LanceDB episode rows can overwrite each other across projects.

The episode primary key is currently constructed as:

```python
id=f"{owner_id}_{entry.entry_id}"

Because app_id and project_id are not included, the same owner’s first daily episode in two different projects receives the same LanceDB ID.
Environment
EverOS: 1.1.0
Python: 3.12
Platform: macOS
Storage: local Markdown + LanceDB
API: /api/v1/memory/add, /flush, /search, /get
All test content and identifiers below are fabricated.
Reproduction
Use the same:
app_id
user_id
date
But two different project IDs:
project-a-test
project-b-test
Add and flush a fabricated memory under Project A.
Confirm that Project A search returns the extracted episode.
Add and flush another fabricated memory under Project B.
Search or list Project A again.
Inspect the Markdown files and LanceDB episode table.
Actual behavior
Both Markdown files exist:
<root>/<app_id>/project-a-test/users/<owner>/episodes/...
<root>/<app_id>/project-b-test/users/<owner>/episodes/...
Both independently allocate:
entry_id = ep_<date>_00000001
They therefore produce the same LanceDB primary key:
<owner_id>_ep_<date>_00000001
After Project B is indexed:
/memory/get for Project A returns 0 episodes.
/memory/get for Project B returns 1 episode.
Project A’s Markdown remains intact.
Direct LanceDB inspection finds only one row for the colliding ID, pointing to Project B.
Project A was searchable before Project B was indexed, confirming that the indexed row was replaced rather than never created.
Expected behavior
Project A and Project B episodes should coexist in LanceDB.
Writing memory under one project_id must not replace an episode belonging to another project_id.
Both of these queries should remain valid:
project_id = project-a-test
project_id = project-b-test
Suspected root cause
EpisodeHandler constructs the row ID without application or project scope:
id=f"{owner_id}_{entry.entry_id}"
The LanceDB repository then performs an upsert keyed by id.
The project scope exists as row fields and query filters, but it is not part of the upsert identity.
Similar ID construction may also exist in other daily-log tables, so atomic_fact, agent_case, and foresight may need review as well.
Suggested direction
Use a deterministic scoped row identity that incorporates at least:
app_id
project_id
owner_id
entry_id
The fix should also include:
Cross-project collision regression tests.
A documented LanceDB rebuild/migration path.
Review of dependent episode/fact references.
Verification that Markdown remains the source of truth.
I can provide additional sanitized probe output if needed.

### Steps to reproduce

1. Start EverOS 1.1.0 with a clean local memory root.

2. Use the same `app_id` and `user_id`, but two different project IDs:

   - `project-a-test`
   - `project-b-test`

3. Add and flush one fabricated memory under `project-a-test`.

4. Confirm `/api/v1/memory/search` or `/api/v1/memory/get` returns the Project A episode.

5. Add and flush another fabricated memory under `project-b-test` on the same day.

6. Query Project A again.

7. Inspect both project Markdown directories. Both episode files exist and both use the same first daily entry ID:

   `ep_<date>_00000001`

8. Inspect the LanceDB `episode` table. Only one row exists for:

   `<owner_id>_ep_<date>_00000001`

   The row now points to Project B.

9. Observe that Project A Markdown still exists, but Project A search/get returns no episode.

### Environment

_No response_

### Logs or screenshots

```shell

```

alexmat:
Fixes #320.

## Problem

Every daily-log cascade handler built LanceDB row ids as `f"{owner_id}_{entry_id}"`, and `user_profile` as bare `owner_id`. Entry numbering is per-file (`<kind>_<date>_NNNNNNNN`), so any two projects with a same-date file produce **identical row ids** — and `merge_insert("id")` treats the previously indexed project's rows as matches and replaces them wholesale. Only the last-processed file per date survives.

Confirmed in production on 1.2.3 (current PyPI latest; `main` was still building ids the same way before this patch), in a multi-project deployment (~20 projects, one owner):

- The `atomic_fact` table stayed pinned at exactly 5,194 rows for 13+ hours while hundreds of commits landed — every insert was silently a replacement.
- Version-walk (`list_versions` + per-version counts) shows the mechanism in lockstep. One transition as the smoking gun: **v2263 → v2264: −54 rows from project `comms`' `atomic_fact-2026-08-27.md`, +54 rows from project `dsh-69826`'s same-date file, with identical entry_ids** (`af_20260827_00000018`, `…46`, `…19`, …).
- The same scheme in `user_profile` (`id = owner_id`) collapsed 16 project profiles into 2 rows — the last two projects written.

Affected handlers: `atomic_fact`, `episode`, `foresight`, `agent_case` (all `id=f"{owner_id}_{entry.entry_id}"`), plus `user_profile` (`id=owner_id`).

## Fix

Row id derives from the file's relative path plus entry id rather than enumerating scope keys — the path is unique by construction and stays correct if a new layout dimension appears later:

- Four daily-log kinds: `id = f"{md_path}#{entry_id}"`
- `user_profile`: `id = md_path` (of the `user.md`), with the recall-side profile fetch deriving the same path from `app_id`/`project_id`/`owner_id` (`ProfileRecaller.fetch` gains `app_id`/`project_id` params; call site in `search/manager.py` passes them from the request)

## Regression test

`tests/unit/test_memory/test_cascade/test_handler_row_id_project_scoping.py` — parametrized over the four daily-log kinds: writes the same-date entry for two different projects through each writer + handler pair against an in-memory repo with merge-by-id semantics, and asserts both projects' rows coexist under distinct ids. **Fails on ≤ 1.2.3 (4/4 fail), passes with this patch (4/4 pass).**

## Migration note for existing deployments

A full `cascade rebuild` is required after upgrading — sha-skip means old-id rows are never rewritten, and new-id rows would otherwise coexist as duplicates. Markdown is canonical, so the rebuild is lossless; parking the optimize/prune loop during the rebuild is advisable. Post-fix verification on our deployment: 14,735 `atomic_fact` rows indexed (2.8× the collision-era steady state), 17 `user_profile` rows, zero cross-project replacements across 168 version transitions and a full optimize+prune cycle.

## Follow-up (not in scope)

`atomic_fact.parent_id` stores the episode's `entry_id`, so fact→episode parent resolution has the same cross-project ambiguity on same-date episode files. Worth its own issue/PR.

### DIFF
--- a/src/everos/memory/cascade/handlers/agent_case.py
+++ b/src/everos/memory/cascade/handlers/agent_case.py
@@ -77,7 +77,7 @@ class AgentCaseHandler(BaseDailyLogHandler):
                 reason="embedding_capability_unavailable",
             )
         return AgentCase(
-            id=f"{owner_id}_{entry.entry_id}",
+            id=f"{md_path}#{entry.entry_id}",
             entry_id=entry.entry_id,
             owner_id=owner_id,
             owner_type=owner_type,
--- a/src/everos/memory/cascade/handlers/atomic_fact.py
+++ b/src/everos/memory/cascade/handlers/atomic_fact.py
@@ -62,7 +62,7 @@ class AtomicFactHandler(BaseDailyLogHandler):
                 reason="embedding_capability_unavailable",
             )
         return AtomicFact(
-            id=f"{owner_id}_{entry.entry_id}",
+            id=f"{md_path}#{entry.entry_id}",
             entry_id=entry.entry_id,
             owner_id=owner_id,
             owner_type=owner_type,
--- a/src/everos/memory/cascade/handlers/episode.py
+++ b/src/everos/memory/cascade/handlers/episode.py
@@ -102,7 +102,7 @@ class EpisodeHandler(BaseDailyLogHandler):
         tokens = self._deps.tokenizer.tokenize(tokenize_source)
 
         return Episode(
-            id=f"{owner_id}_{entry.entry_id}",
+            id=f"{md_path}#{entry.entry_id}",
             entry_id=entry.entry_id,
             owner_id=owner_id,
             owner_type=owner_type,
--- a/src/everos/memory/cascade/handlers/foresight.py
+++ b/src/everos/memory/cascade/handlers/foresight.py
@@ -86,7 +86,7 @@ class ForesightHandler(BaseDailyLogHandler):
             " ".join(self._deps.tokenizer.tokenize(evidence)) if evidence else None
         )
         return Foresight(
-            id=f"{owner_id}_{entry.entry_id}",
+            id=f"{md_path}#{entry.entry_id}",
             entry_id=entry.entry_id,
             owner_id=owner_id,
             owner_type=owner_type,
--- a/src/everos/memory/cascade/handlers/user_profile.py
+++ b/src/everos/memory/cascade/handlers/user_profile.py
@@ -80,7 +80,7 @@ class UserProfileHandler(Handler):
             }
         )
 
-        row_id = owner_id
+        row_id = md_path
         prior = await user_profile_repo.get_by_id(row_id)
         if prior is not None and prior.content_sha256 == digest:
             return HandlerOutcome(
--- a/src/everos/memory/search/manager.py
+++ b/src/everos/memory/search/manager.py
@@ -576,7 +576,7 @@ class SearchManager:
     async def _fetch_profile(self, req: SearchRequest) -> list[SearchProfileItem]:
         if not req.include_profile or req.owner_type != "user":
             return []
-        return await self._profile.fetch(req.owner_id)
+        return await self._profile.fetch(req.owner_id, req.app_id, req.project_id)
 
     # ── Recall helpers ──────────────────────────────────────────────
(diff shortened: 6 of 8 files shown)


BrierAinz:
This is a strong fix direction. Scoping row ids by canonical markdown path instead of enumerating known dimensions is exactly the defensive move for future layout keys.

Two migration details seem worth making explicit before release:

1. The rebuild should be documented as mandatory, not best-effort, because old-id rows and new-id rows can coexist after the first post-upgrade cascade.
2. The profile recall path now derives `user.md` from `app_id` / `project_id` / `owner_id`; I would include a negative regression case proving two projects with the same owner return different profiles through the public search path, not only through handler-level row creation.

That second check closes the loop on the riskiest part of the change: writes are now path-scoped, but reads must be path-scoped with the exact same canonicalization.

Issue #443, opened 2026-09-09
[Bug] Reflection V1: large single-cluster merges cause partial deprecated_by writes with no compensation, no per-merge source cap, and no prompt/language configuration


ProjectAILeap:
## Summary

Three related defects in Reflection V1 (`reflect_episodes`) that compound on large clusters:

1. **Partial deprecation with no compensation** — the per-episode LanceDB `deprecated_by` updates run as N concurrent writes; when some exceed the write-lock deadline under load, the run is reported failed while the already-applied updates persist. Sources are left pointing `deprecated_by` at a merged entry that may be invalid or already removed, with no rollback or reconcile pass.
2. **No per-cluster source cap** — a merge consumes *all* members of a cluster (observed `source_count=228` → one narrative). Only `_MAX_CLUSTERS_PER_RUN = 10` is bounded. Oversized merges produce narratives whose cascade upsert exceeds the 15 s write-lock deadline, starving the same run's deprecate updates (feedback loop, see below).
3. **No prompt/language entry point** — the reflector prompt is hard-wired to the upstream English prompt; merged output language is left to the model (observed mixed Chinese/English/French, occasionally garbage), with no `[reflection]` config to inject instructions.

## Environment

- EverOS `1.3.1`, everalgo-core `0.3.0`, LanceDB index
- LLM: `mimo-v2.5` via an OpenAI-compatible proxy that *does* support strict structured outputs — the merge LLM call succeeds, so this is not a provider-shape failure

## Observed behavior

Manual trigger over an existing 228-member cluster:

```
reflection_merged ... cluster_id=cl_bb3f91313016 mode=init source_count=228
lancedb_write_lock_deadline_exceeded ... op=update table=episode budget_seconds=15.0 acquired=False (repeated xN)
reflection_deprecate_failed ... cluster_id=cl_bb3f91313016
```

After `reflection_deprecate_failed`, LanceDB inspection showed **179–297 `episode` rows (and thousands of `atomic_fact` rows) with `deprecated_by` already written** pointing at the merged entry — deprecation partially applied before failing, and nothing reconciles the leftovers. Concurrently `cascade_worker_recoverable` logged `VectorStoreBusyError: upsert on table 'episode' exceeded its 15s write-lock deadline` while indexing the same oversized narrative.

## Root cause

**Timeout mechanism (measured).** The storm is queue-wait starvation behind the single per-table lock, not slow individual ops:

[...]

PR #445, opened 2026-09-09
fix(reflection): batch source deprecation, revert it on failure, cap sources per merge

L4XB:
## Summary

Addresses the first two defects in #443 (Reflection V1 on large clusters):

**Why the run failed.** `_deprecate_lance_episodes` / `_deprecate_lance_facts` issued one LanceDB `update` per source episode and per fact parent, all started concurrently with `asyncio.gather`. Every `update` takes the table's write lock, so on a 228-member cluster the calls queued behind one another until the 15 s write-lock deadline (`lancedb_write_lock_deadline_exceeded` storm), the run was reported failed, and the updates that had already gone through stayed applied: sources half-deprecated toward a merge that was never committed to the cluster.

**Changes**

1. **Batched, sequential deprecation.** Sources are deprecated in batches of 100 ids with one `entry_id IN (...)` (episodes) / `parent_id IN (...)` (facts) update each, issued sequentially. A merge now takes a handful of lock acquisitions instead of hundreds.
2. **Compensation on failure.** The LanceDB rows are written first and the markdown frontmatter last, so the durable record only lists deprecations that were actually applied. If any write fails, the `deprecated_by` values written by this run (only rows pointing at this merge's entry id) are cleared before the error propagates. A failed run now leaves the sources, the markdown record and the cluster as they were; the merged episode itself is still detected by the existing orphan check on the next run, as before.
3. **Per-merge source cap.** `_MAX_SOURCES_PER_MERGE = 50` (module constant next to `_MAX_CLUSTERS_PER_RUN`). A larger cluster merges its existing merged episode(s) plus the oldest sources up to the cap; deferred members stay in the cluster, are neither merged nor deprecated in this run, and are folded in by later runs in update mode. The cluster's `count` after a merge now reflects the members that remain (previously hard-coded to 1). Wiring the cap into `[reflection]` config can follow if you prefer it configurable.

The third point in the issue (prompt / output-language configuration) is a product decision and is left out.

## Area

- [x] Architecture method

## Verification

```text
uv run pytest tests/unit/test_memory          711 passed
make lint                                     ruff, import-linter, datetime discipline, openapi drift: OK
```

New tests in `tests/unit/test_memory/test_reflection/test_orchestrator.py`:

- `test_deprecate_lance_episodes_batches_updates`: 250 ids → 3 updates with `IN` predicates covering every id.
- `test_deprecate_failure_reverts_applied_writes`: the second batch raises `VectorStoreBusyError` → the markdown record is never patched, both applied batches are reverted (`deprecated_by = NULL` only where it equals this merge's entry id), and the error propagates so the run is still reported failed.
- `test_run_caps_sources_per_merge_and_keeps_the_rest`: with the cap at 2 and a 3-member cluster, only the two oldest sources are reflected, deprecated and removed; the deferred member plus the merged episode remain (`count=2`).

## Checklist

## Notes for Reviewers

The reorder (LanceDB first, markdown last) is what makes the compensation complete without having to un-patch frontmatter. If you would rather keep markdown first, the revert still works for the LanceDB side, but the frontmatter would then re-apply the deprecation on the next cascade reconcile.

Refs #443

### DIFF
--- a/src/everos/memory/reflection/orchestrator.py
+++ b/src/everos/memory/reflection/orchestrator.py
@@ -10,7 +10,6 @@ See ``local/2026-06-14-reflection-everos-design.md`` for the full design.
 
 from __future__ import annotations
 
-import asyncio
 import datetime as _dt
 import json
 import uuid
@@ -32,13 +31,25 @@ from everos.core.observability.logging import get_logger
 from everos.core.observability.tracing import memory_span
 from everos.core.persistence import MemoryRoot
 from everos.infra.ome.context import StrategyContext
-from everos.infra.persistence.index import all_of, eq, is_null
+from everos.infra.persistence.index import all_of, eq, is_null, one_of
 from everos.memory._partition_locks import get_partition_lock
 from everos.memory.events import EpisodeExtracted
 
 logger = get_logger(__name__)
 
 _MAX_CLUSTERS_PER_RUN = 10
+
+# Upper bound on the episodes one merge consumes. A cluster with more members
+# is merged incrementally: its existing merged episode plus the oldest sources
+# go into this run, the remaining members stay in the cluster and are folded
+# in by later runs in update mode. Keeps the merged narrative, and the
+# cascade upsert it triggers, inside the LanceDB write-lock budget.
+_MAX_SOURCES_PER_MERGE = 50
+
+# Entry ids per LanceDB ``update`` when deprecating sources. One update per
+# batch instead of one per row keeps the number of write-lock acquisitions
+# small for large clusters.
+_DEPRECATE_BATCH_SIZE = 100
 _WAIT_TIMEOUT_SECONDS = 120.0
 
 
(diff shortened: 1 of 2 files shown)


BrierAinz:
The LanceDB-first / markdown-last ordering is a good fix for the half-deprecated failure mode. I like that the compensation predicate clears only rows pointing at this merge entry id; that avoids undoing unrelated deprecations from another successful run.

One edge I would want covered, either by test or review note: retry after a compensated failure should be idempotent when the failed run created the merged episode row but did not patch markdown frontmatter. In other words, the orphan detection path should either remove/ignore that merge candidate before the retry or guarantee it cannot be selected as a source and produce a duplicate reflection.

The source cap also changes operator expectations a bit: a large cluster is now intentionally convergent across multiple runs, not atomic in one run. If there is run telemetry, exposing `sources_deferred` / `cluster_remaining_count` would make this much easier to distinguish from a silently incomplete reflection.

L4XB:
@BrierAinz Good catch. The edge was real, so I fixed it rather than just adding a note.

**Retry after a compensated failure.** The failed run's merged episode stays live and outside the cluster. It can never be selected as a source, because sources come from cluster membership and the membership update is the step that never ran. But the retry wrote a second merged episode beside it. A test that makes the first deprecation write miss its lock deadline, lets compensation revert it, and then retries ended with two live merged episodes at the previous head:

```
assert [r.entry_id for r in live] == [merged_id]
AssertionError: assert ['ep_20260923_00000004', 'ep_20260923_00000005'] == ['ep_20260923_00000005']
```

9f6d620 moves orphan detection from the start of the run, where it only logged, into the deprecation step. Live merged episodes of the cluster that are neither current members nor this run's merge are deprecated toward the new merge in the same writes, and reverted with them if those fail, so a retry converges on one live merged episode. The orphan's markdown is patched like any source's: a merged episode is written to markdown first, so it always has an `md_path`. `test_retry_after_failed_deprecation_leaves_one_merged_episode` passes for both the lance and markdown failure points and fails without the change. The integration and orchestrator unit files pass (18 tests), and `make lint`'s gates are clean.

**Telemetry for the cap.** A capped run already logs `reflection_cluster_capped` with `deferred_count`. f03d7f5 adds `cluster_remaining_count` (the members this merge left for a later run) and `retired_orphan_count` to `reflection_deprecated`, so a cluster that converges over several runs is visible as such.

PR #451, opened 2026-09-16
fix(search): apply cosine radius across iterative recall paths

### LINKED ISSUE #394: radius is documented and validated but has no effect on any search lane
`SearchRequest.radius` is declared, range-validated to `[0.0, 1.0]`, and documented at `memory/search/dto.py:84-88` as gating "raw cosine at recall time". It has no effect anywhere.

- **HYBRID**: `manager.py` passes `radius=_effective_radius(req)` into `RankInput` at three call sites (~`:420`, `:477`, `:551`). `grep -rn radius .venv/.../everalgo/` returns exactly two hits, both in `types/rank.py` — the field declaration and its docstring. everalgo never reads it.
- **AGENTIC**: never passed at all.

PR #393 briefly wired it into the agentic path as `ahybrid_retrieve(min_score=radius)` and had to revert: `min_score` is applied *after* RRF fusion (`everalgo/rank/hybrid.py:86-91`), where `rrf` replaces each score with `Σ 1/(k+rank)` (max ≈ 0.0328, `fusion.py:41`). Since `top_k` defaults to `-1` and `_effective_radius` then returns `_DEFAULT_UNLIMITED_RADIUS = 0.5` (`manager.py:120`, `:937-941`), every candidate was filtered out — default agentic agent search returned `[]`. Measured: `{None: 2 hits, 0.5: 0 hits, 0.03: 2 hits}`.

Needs a decision before any implementation: what should `radius` mean? Candidates — (a) a pre-fusion cosine floor applied inside the dense recall route, which is what the docstring describes; (b) a post-fusion floor on the calibrated LR score, which is what `min_score` already does; (c) remove it and keep `min_score`. Whichever is chosen, apply it consistently across lanes and delete the three inert `RankInput.radius` args.

Found while verifying #393.

knqiufan:
## Summary

Fixes #394.

`agentic` and `llm_multiround` accepted `radius` but bypassed the manager's dense-recall filter. A request with dense hits at cosine 0.9 and 0.2 and `radius=0.5` now removes the 0.2 hit before fusion on every iterative query. Existing `vector` and `hybrid` filtering remains in place.

- Share request-default resolution and dense filtering across search orchestration. Resolve the default before capping `top_k`: explicit values win (including zero), unlimited searches default to 0.5, and positive `top_k` has no implicit floor.
- Pass the resolved floor through user/agent AGENTIC and LLM_MULTIROUND. Filter both atomic-fact and episode-subject dense hits before user AGENTIC MaxSim pooling, and filter each multi-round dense pool before RRF.
- Remove the three inert `RankInput.radius` arguments. Never map cosine thresholds to `ahybrid_retrieve(min_score=...)`, which would compare them with a different score scale.
- Document the dense-only contract and the episode HYBRID-only scope of `min_score`. Add 49 regression cases covering dispatch/defaults, boundaries, sparse preservation, empty dense pools, both agent kinds, and repeated queries with real RRF/MaxSim operators.

## Area

- [x] Architecture method
- [ ] Benchmark
- [ ] Use case
- [x] Documentation
- [ ] Developer experience
- [ ] CI, build, or release

## Verification

Pre-PR Linux validation uses the same source/test tree as this PR; validation branches change only workflow configuration:

- [Linux CI](https://github.com/knqiufan/EverOS/actions/runs/35064634620): `make lint` and `make package` passed; `make test` passed on Python 3.12 and 3.13 (2599 passed, 4 skipped each); `make integration` passed on both versions (183 passed, 5 skipped, 7 deselected each).
- Local search suite: 310 passed. Windows required an import-only `fcntl` stand-in that raises on any attempted POSIX operation; the Linux runs above use the real platform without that stand-in.
- Documentation links, Conventional Commit format, and `git diff --check` passed. OpenAPI export has no drift.
- [Milvus 2.6.22 validation](https://github.com/knqiufan/EverOS/actions/runs/35064785446): repository contract (5 passed), `/get` telemetry (1 passed), tiered API end-to-end tests (27 passed), and both-backend index contract (18 passed). The first run could not pull the upstream compose file's `minio/minio` image; the successful validation uses the same MinIO version tag from `quay.io/minio/minio`. This environment-only workaround is not part of the PR; the unchanged upstream workflow may encounter the same Docker Hub pull failure.

## Checklist

## Notes for Reviewers

This implements option (a) from the issue: a pre-fusion dense cosine floor. Keyword matches, cluster expansions, and case-to-skill linkage remain independent recall routes; radius is not a final-score floor or a promise of monotonically decreasing result counts. The existing `min_score` behavior is unchanged.

The intended behavior change includes default unlimited iterative searches: they now honor the documented 0.5 floor. Internal helper callers that omit the new optional argument retain unfiltered behavior. VECTOR and HYBRID already filtered dense hits before this change; the unused ranker argument was not evidence that those paths ignored radius.

By submitting this pull request, I agree that my contribution is licensed under the Apache License 2.0.


BrierAinz:
This looks like the right shape for `radius`: resolve once as a dense cosine floor, apply it before fusion, and keep sparse / linkage routes out of that promise.

One acceptance check I would keep explicit before merge: a repeated-query iterative request should prove that each dense pool is filtered independently before RRF/MaxSim, not only that the final result set happens to exclude a low-score item. The easy regression trap is a helper path that filters the first query's pool but lets a later round contribute sub-radius dense hits through fusion.

It may also be worth spelling out in docs that `radius=0` means "accept every dense hit" while omitted unlimited iterative searches default to `0.5`. That distinction is subtle, but it is exactly the kind of subtlety client SDKs tend to flatten accidentally.

